PR #9, opened 2019-09-02
Completed Problem 1.

priyansh22:


sehaj-grover (commented):
very good  but see if you can handle the problem without sorting the array

priyansh22:
Yes I was thinking of another solution that could reduce the time complexity. 
Is the time complexity mentioned by me correct for the code? I am not sure if it should be O(nlogn).

sehaj-grover (commented):
the time complexity for backtracking solution is always exponential

priyansh22:
So according to my code should the time complexity be O(n^2 logn) or O(n^2). Because it involves sorting the elements too.

sehaj-grover (commented):
it's 2^n and this approach will itself work even if you remove the sort function

priyansh22:
Oh yes. It does work. So sorting the list was not required.
Thank you.

PR #13, opened 2019-09-09
Completed Backtracking-1

antopraju:


sehaj-grover (commented):
good
keep submitting the pending work

PR #14, opened 2019-09-23
Please review

Jaysparkexel:


sehaj-grover (commented):
jay try the other problem as well

Jaysparkexel:
@sehaj-grover I tried the problem but wasn't able to solve it.

sehaj-grover:
Okay try again now after the session as see if you can figure it out else just ping me and we'll solve it together

PR #44, opened 2020-01-29
Kindly review backtracking

sachinsharma3191:


rajatkumar01 (commented):
Good work Sachin. You are using two languages for solving problems. Is this intentionally or copying it from some where....

sachinsharma3191:
Python code is more readable.Being a dynamic language ,it helps avoid writing more code.Like in case of largest value in binary tree, i have to type more code in java or c++.In python its written in less lines of code

rajatkumar01:
:)

PR #74, opened 2020-04-20
Backtrackign-1

AishwaryaSharma27:
Please review

### DIFF
--- a/CombinationSum.java
+++ b/CombinationSum.java
@@ -1,0 +1,27 @@
+//Time complexity: O(n!)
+//Space complexity: O(1)
+
+import java.util.*;
+
+class CombinationSumSolution {
+    public List<List<Integer>> combinationSum(int[] candidates, int target) {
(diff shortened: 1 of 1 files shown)


CombinationSum.java
```diff
@@ -0,0 +1,27 @@
+//Time complexity: O(n!)
```
devashishnyati, line 1: My understanding of the time complexity is that it will be O(2^n). It's better to say that the complexities will be exponential.

CombinationSum.java
```diff
@@ -0,0 +1,27 @@
...
+//Space complexity: O(1)
```
devashishnyati, line 2: Recursion stack space?

devashishnyati (commented):
Good job Aishwarya!
Submit the other problem as well after the class.

devashishnyati:
Good job Aishwarya!
Submit the other problem as well after the class.

PR #86, opened 2020-05-26
Completed Backtracking-1: Please Review

letscodeitout17:


CombinationSum.java
```diff
@@ -0,0 +1,66 @@
+// Time Complexity : O(2^n) where n is hte number of candidates 
+// Space Complexity : O(target/min(n)) where n is the number of candidates
```
super30admin, line 2: Exponential in recursive.

OperatorComb.java
```diff
@@ -0,0 +1,93 @@
+// Time Complexity : O(4^n) where n is the number of elements in the array and we have 4 decisions at each instance 
+// Space Complexity : O(n) where n is the number of operands in the stringbuilder
```
super30admin, line 2: Exponential in case of recursive

super30admin (commented):
Good Job Anik. Please review inline comments

CombinationSum.java
letscodeitout17: Oh, this is for backtracking.

OperatorComb.java
letscodeitout17: Oh, this is for backtracking.

letscodeitout17: @super30admin I have a question here:
In Expression Add, We have 0 as a special case "if(index != i and num.charAt(index) == 0) break".
In the code, I couldn't figure out why the 0 is getting ignored/skipped in the case of "105" as input, because it should take 0 in the number into consideration as per code and how the code line above is fixing it?

super30admin: This is ignoring just the preceding zero **for pivot**.. That is ignore processing the pivot at zero and move to next pivot.

PR #96, opened 2020-05-31
BackTracking-1 done.pls do review

Mohan110594:


AroopN (commented):
Good job Mohan.
1. expression-add-operators
Use list to store the temporary expression in the backtrack function so that insertion and removal can be done in O(1) time. You can use join() to convert the list to string while adding to result.

Please check the space complexity to include only the auxiliary space used.

2. Combination sum

Instead of calculating the sum every time, we can reduce the time by just passing the sum. along with the temporary list.

Please check the time and space complexity.

AroopN:
Good job Mohan.

expression-add-operators
Use list to store the temporary expression in the backtrack function so that insertion and removal can be done in O(1) time. You can use join() to convert the list to string while adding to result.
Please check the space complexity to include only the auxiliary space used.

Combination sum
Instead of calculating the sum every time, we can reduce the time by just passing the sum. along with the temporary list.

Please check the time and space complexity.

PR #124, opened 2020-07-09
Completed both problems, please review.

samarthjj:


AbhiniveshP:
Nice work Samarth!!!
Do you know how to avoid usage of global variables like you've done for backtracking solutions in python?
Do you know what all data types in python are passed by reference by default in python? and what data types can't be passed by reference and can be passed by value only? This particular concept is applicable in other languages as well...

AbhiniveshP:
Nice work Samarth!!!
Do you know how to avoid usage of global variables like you've done for backtracking solutions in python?
Do you know what all data types in python are passed by reference by default in python? and what data types can't be passed by reference and can be passed by value only? This particular concept is applicable in other languages as well...

combinationSum.py
```diff
@@ -0,0 +1,74 @@
...
+        if sum > self.target or idx == len(self.cand):
+            return
+
+        if sum == self.target:
+            self.retVal.append(temp)
+            return
+
+        # case not choose
+        self.helper(idx + 1, sum, list(temp))
+        # case choose
```
samarthjj, line 33: WOJFIFJS

PR #195, opened 2020-11-18
Completed both the problems

kateganesh:


samarthjj:
Good work Ganesh.

samarthjj:
Good work Ganesh.

PR #270, opened 2021-02-23
bc 1

HarshaVardhan-Kaki:


nsibal (commented):
Try to visualize the problem as a tree and find how many nodes will be there in the tree.
For eg in Combination Sum there will at most N nodes and each node will have at most N subtrees. So, the TC would be O(N^N).
At the most, the recursive stack will go deepest in the tree. So, the SC would be O(N).

nsibal:
Try to visualize the problem as a tree and find how many nodes will be there in the tree.
For eg in Combination Sum there will at most N nodes and each node will have at most N subtrees. So, the TC would be O(N^N).
At the most, the recursive stack will go deepest in the tree. So, the SC would be O(N).

PR #328, opened 2021-06-29
I hate humanity for coming up with this kind of stuff. Please review!

kgajwan1:


keshav-Gupta-IS (commented):
Good Work, Karan!! The TC for 1st question is O(2^n) and for the 2nd problem is O(n*4^n). If you want to clarify how we arrived at the TC you can ping me on slack.

keshav-Gupta-IS:
Good Work, Karan!! The TC for 1st question is O(2^n) and for the 2nd problem is O(n*4^n). If you want to clarify how we arrived at the TC you can ping me on slack.

PR #354, opened 2021-07-06
Completed

s4git21:


Problem-1.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+
+
+
+
+
+
+
+
```
rimzimt, line 174: Excellent work with al the approaches!

Problem-2.py
```diff
@@ -0,0 +1,90 @@
...
+                self.helper(num, i + 1, path, calc - int(curr), -1 * int(curr))
+                path.pop()
+                path.pop()
+
+                # * operator
+                path.append('*')  # action
+                path.append(curr)  # action
+                self.helper(num, i + 1, path, calc - tail + tail * int(curr), tail * int(curr))
+                path.pop()
+                path.pop()
```
rimzimt, line 90: Keep it up!

rimzimt (commented):
Good work Syed!

rimzimt:
Good work Syed!

PR #428, opened 2021-10-29
Backtracking-1 completed please review

srinathsai:
Can you tell me time complexity of second sum?

### DIFF
--- a/combination sums.java
+++ b/combination sums.java
@@ -1,0 +1,31 @@
+//Time complexity:- O(N^target/min);(N^height of tree in which height of tree will be untill target/min.
+//space complexity:- O(target/min) no of nodes.
+//Did it run on leetcode?:-yes.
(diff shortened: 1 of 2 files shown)


combination sums.java
```diff
@@ -0,0 +1,31 @@
+//Time complexity:- O(N^target/min);(N^height of tree in which height of tree will be untill target/min.
```
manojvarmarudraraju, line 1: Capital O is worst case, as we are iterating through the array for every element N power N will be worst case time complexity

operators.java
```diff
@@ -0,0 +1,43 @@
+//Time complexity:-
```
manojvarmarudraraju, line 1: Ur doing four operation at every step so it will be 4 power N

manojvarmarudraraju (commented):
Good job!!

manojvarmarudraraju:
Good job!!

PR #429, opened 2021-10-30
Complete

hemanthatushara:


Problem1.py
```diff
@@ -0,0 +1,27 @@
+#Time Complexity : O(2^t) where t is the target value
```
manojvarmarudraraju, line 1: Check the time complexity, 2^n because we are iterating array. So worst case we have to touch every element.

Problem2.py
```diff
@@ -0,0 +1,20 @@
+class Solution:
```
manojvarmarudraraju, line 1: Please add the time and space complexities

manojvarmarudraraju (commented):
Good job, please add the complexities

manojvarmarudraraju:
Good job, please add the complexities

PR #538, opened 2022-05-23
Completed BT - 1. Please review

Meena-hash:


Sum.py
```diff
@@ -0,0 +1,28 @@
+# // Time Complexity : Exponential
```
Sumesh-H, line 1: Please reconsider time and space complexity. O(N*2^N)My understanding is that N is the size of the candidate array. N is the decision space and 2^N is exploring each decision space's different combinations.

Sumesh-H (commented):
Good Job Meena, please address the comment.

Sumesh-H:
Good Job Meena, please address the comment.

PR #604, opened 2022-07-30
Completed Sprint Backtracking-1 : Please Review

SreevikramChandrasekhar:


Problem2_1.java
```diff
@@ -0,0 +1,54 @@
+//Time Complexity: O(2^n)
```
ishika2741, line 1: It will be exponential but more than 2^n

ishika2741 (commented):
Well done!

ishika2741:
Well done!

PR #652, opened 2022-10-17
combination sum

kundankarthik77:
# Bactracking-1

## Problem1 
Combination Sum (https://leetcode.com/problems/combination-sum/)

Given a set of candidate numbers (candidates) (without duplicates) and a target number (target), find all unique combinations in candidates where the candidate numbers sums to target.

The same repeated number may be chosen from candidates unlimited number of times.

Note:

All numbers (including target) will be positive integers.
The solution set must not contain duplicate combinations.
Example 1:

Input: candidates = [2,3,6,7], target = 7,
A solution set is:
[
  [7],
  [2,2,3]
]
Example 2:

Input: candidates = [2,3,5], target = 8,
A solution set is:
[
  [2,2,2,2],
  [2,3,3],
  [3,5]
]

## Problem2
Expression Add Operators(https://leetcode.com/problems/expression-add-operators/)

Given a string that contains only digits 0-9 and a target value, return all possibilities to add binary operators (not unary) +, -, or * between the digits so they evaluate to the target value.

Example 1:

Input: num = "123", target = 6
Output: ["1+2+3", "1*2*3"] 
Example 2:

Input: num = "232", target = 8
Output: ["2*3+2", "2+3*2"]
Example 3:

Input: num = "105", target = 5
Output: ["1*0+5","10-5"]
Example 4:

Input: num = "00", target = 0
Output: ["0+0", "0-0", "0*0"]
Example 5:

Input: num = "3456237490", target = 9191
Output: []

### DIFF
--- a/expressionAddOperations.java
+++ b/expressionAddOperations.java
@@ -1,0 +1,40 @@
+class Solution {
+    List<String> result;
+    public List<String> addOperators(String num, int target) {
+        result = new ArrayList<>();
+        if (num == null || num.length() == 0)
+            return result;
+        helper(num, target, 0, "", 0, 0);
+        return result;
+    }
+    
+    private void helper(String num, int target, int index, String path, long calc, long tail) {
+        // base
+        if (index == num.length()) {
+            if (calc == target) {
+                result.add(path);
+            }
+            return;
+        }
+        // logic
+        for (int i = index; i < num.length(); i++) {
+            if (index != i && num.charAt(index) == '0') continue;
+            long curr = Long.parseLong(num.substring(index, i+1));
+            if (index == 0) {
+                helper(num, target, i + 1, path + curr, curr, curr);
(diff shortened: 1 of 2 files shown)


expressionAddOperations.java
```diff
@@ -0,0 +1,40 @@
...
+                helper(num, target, i + 1, path + "-" + curr, calc-curr, -curr);
+                // * operator case
+                helper(num, target, i + 1, path + "*" + curr, calc-tail + tail * curr, tail * curr);
+            }
+        }
+    }
+}
+
+// Executed
+// T.C : O(4 ^ n) where 4 is number of operations
```
daredevilteja, line 39: TC is O(n*4^n) and SC is O(n)

daredevilteja:
Good Work! Keep it up

PR #1090, opened 2025-09-26
Add solutions for Combination Sum and Expression Add Operators

ravithakur2k:
Added CombinationSum.py with recursive and iterative backtracking solutions for the Combination Sum problem. Added ExpressionAddOperators.py with two recursive solutions for generating expressions that evaluate to a target value, including optimizations with backtracking.

### DIFF
--- a/CombinationSum.py
+++ b/CombinationSum.py
@@ -1,0 +1,59 @@
+# Time Complexity : O(2^(m+n)) where m is the length of candidates and n is the target
+# Space Complexity : O(n)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- The student demonstrates a strong understanding of backtracking and recursion.
- The student provides multiple approaches (recursive and iterative) for the Combination Sum problem, showing versatility.

Areas for Improvement:
- In the CombinationSum.py file, the student has defined two `Solution` classes, which is not standard practice. It would be better to have one class with both methods or separate the solutions into different files.
- For the Expression Add Operators problem, the student could consider adding more comments to explain the `calc` and `tail` variables, as their roles might not be immediately obvious to someone reading the code for the first time.
- The student could also consider adding more test cases or edge cases in the comments to demonstrate the robustness of the solution.

PR #1126, opened 2026-02-22
Backtracking 1

eyeamkd:
This pull request adds two new solution implementations for algorithmic problems, each with detailed time and space complexity analyses in docstrings. The solutions use recursive backtracking to solve combination and expression generation problems.

**New solution implementations:**

* Combination Sum:
  - Implements a recursive backtracking approach in `combination-sum.py` to find all unique combinations of candidate numbers that sum to a target. Includes a detailed explanation of time and space complexity.

* Expression Operators:
  - Implements a recursive solution in `expression-operators.py` to add operators between digits to reach a target value, with a comprehensive complexity analysis.


super30admin:
For the "Combination Sum" solution:
- Well done! Your solution is correct and efficient. The backtracking approach with the `index` parameter to avoid duplicates is implemented correctly. The time and space complexity analysis is accurate.

For the "Expression Add Operators" solution:
- There is a flaw in handling multiplication. The common approach to handle operator precedence is to keep track of the last operand (prev_value) so that when multiplication is encountered, you can undo the last operation and then multiply. However, your formula `(current_value - prev_value) + (prev_value * value)` is correct only if the last operation was addition or subtraction. Actually, it should work for multiplication as well because:
  - For addition: current_value = current_value + value, and prev_value = value.
  - For subtraction: current_value = current_value - value, and prev_value = -value.
  - For multiplication: current_value = (current_value - prev_value) + (prev_value * value), and prev_value = prev_value * value.
This method is standard and should be correct. But let me test with an example: 
  Suppose we have "2+3*4". 
  Steps: 
    Start: current=0, prev=0.
    Then we take "2": current=2, prev=2.
    Then we add "+3": current=2+3=5, prev=3.
    Then we add "*4": current = (5 - 3) + (3 * 4) = 2 + 12 = 14, which is correct.
So the multiplication formula is actually correct. 
However, there is another issue: the code does not account for the fact that numbers can be multi-digit, but that is handled by the for loop. 
But wait, the problem requires that we output all expressions that evaluate to the target. Your solution might be correct in terms of calculation, but I notice that you are not passing the expression string correctly. In the multiplication case, you are building the expression with the operator and the number, which is correct.

However, there is a known issue: when the number is zero, we should not generate numbers with leading zeros. Your check `if num[index]=='0' and i>index: break` is correct.

But why did I think it was incorrect? Actually, upon second look, the multiplication handling is standard and correct. So the solution for "Expression Add Operators" might be correct.

But wait, there is a potential problem: the initial call is `solve(0, "", 0, 0)`, which is correct. However, in the recursive calls for non-initial indices, you are building the expression by concatenating the operator and the number. This is correct.

So both solutions appear to be correct. However, for "Expression Add Operators", the time complexity might be higher because of the string concatenation. Each string concatenation creates a new string, which takes O(length of expression) time. The total time complexity could be O(N * 4^N) because each expression has length O(N).

Also, note that the problem "Expression Add Operators" is more complex and has edge cases (like leading zeros). Your solution handles leading zeros correctly.

Therefore, both solutions are correct.

Overall, the student has done a good job. The code is clean and efficient. The complexity analysis is reasonable.

PR #1133, opened 2026-04-14
Backtracking-1

paridhimalviya:
Expression Add Operators(https://leetcode.com/problems/expression-add-operators/)
Combination Sum (https://leetcode.com/problems/combination-sum/)


super30admin:
### Combination Sum (CombinationSum.swift)
Your solutions are correct and well-implemented. Here are some points to consider:
1. The "0-1 recursion" solution is correct, but in the no-choose branch, you don't need to create a new copy of the path since you are not modifying it. You can simply pass the same path array. This will save memory because Swift uses copy-on-write for arrays. However, in the choose branch, you are creating a new copy by using `var path = path` and then appending. This is acceptable, but the for-loop based solution uses backtracking which is more memory efficient.
2. The for-loop based solution is excellent because it uses backtracking to avoid creating multiple copies. This is the preferred approach for such problems.
3. You have included a permutation variant which is interesting, but note that the problem requires combinations (order does not matter). So the permutation solution would generate duplicates if the same numbers are in different orders. However, since you are passing the start index in the for-loop based recursion, you avoid duplicates and generate only combinations. The permutation solution you provided (getPermutationsWithDifferentArrangements) would actually generate permutations, which is not required. But it's good to have for learning.
4. Your code is well-commented and shows understanding of time and space complexity. Keep up the good work!

VERDICT: PASS

---

### Operations and expressions (ExpressionAddOperators.swift)
Your solution shows a good understanding of the problem and you've implemented two versions: one with immutable paths and one with backtracking. However, there is a critical error in handling numbers with leading zeros. You should break out of the loop when you encounter a number that starts with '0' and has more than one digit (i.e., when `start != i` and the first digit is '0'). Currently, you use `continue`, which skips the current index but continues with larger numbers. This is incorrect because if the substring starts with '0' and has length greater than 1, it is invalid. For example, in "105", the substring "05" is invalid. So you should break instead of continue.

Also, consider using a larger integer type if necessary, but since Swift Int is 64-bit, it should be safe for the constraints.

Your backtracking solution (addOperatorsB) is more efficient because it mutates the path in place. However, you need to fix the leading zero issue there as well.

Another note: in the first solution, you are passing `path + substr` for the first level. This creates a new array each time. This is acceptable for small n (<=10), but it may use more memory. The backtracking solution is better in this regard.

You might also want to consider using a string builder pattern or an array of characters that you mutate and then convert to a string only when you add to the result. This can reduce memory overhead.

Overall, your approach is correct except for the leading zero handling. Fix that and your solution should work.

VERDICT: NEEDS_IMPROVEMENT

